PR #11, opened 2017-04-04
Rename intrumentation artifact to opentracing-spring-web

pavolloffay:


pavolloffay:
> Would also consider moving autoconfigure into the single spring-web artifact unless good reason to keep it separate. 
https://github.com/opentracing-contrib/java-spring-web/issues/8#issuecomment-291424739

pavolloffay:
@objectiser thanks

PR #14, opened 2017-04-10
Skip Pattern to exclude certains URLs

pavolloffay:


pavolloffay:
build passed, and locally it says that `io.opentracing.contrib:opentracing-web-servlet-filter:jar:0.0.6` is not in central yet..

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
```diff
@@ -26,12 +25,19 @@
...
+    public WebTracingConfiguration tracerAutoConfiguration() {
+        return WebTracingConfiguration.builder()
+                .withSkipPattern(WebTracingConfiguration.DEFAULT_SKIP_PATTERN).build();
+    }
+
+    @Bean
+    public FilterRegistrationBean tracingFilter(Tracer tracer, WebTracingConfiguration tracingConfiguration) {
         log.info("Creating " + FilterRegistrationBean.class.getSimpleName() + " bean with " +
-                TracingFilter.class + " mapped to " + "/*");
+                TracingFilter.class + " mapped to " + "/*, skip pattern is " + tracingConfiguration.skipPattern);
```
jpkrohling: You are subject to a NPE here. Perhaps you'd want to do a null-check?

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebTracingConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.regex.Pattern;
+
+/**
+ * @author Pavol Loffay
+ */
+public class WebTracingConfiguration {
+    public static final Pattern DEFAULT_SKIP_PATTERN = Pattern.compile("/health|/favicon\\.ico");
+
+    public final Pattern skipPattern;
```
jpkrohling: Copy/paste what I said on IRC about this field being public:

> I have nothing against it per se, *but*, it's a common style to provide getters/setters to properties in Java. If you are deviating from what people expect, make sure to have a reason and document it

opentracing-spring-web-itest/common/src/test/java/io/opentracing/contrib/spring/web/interceptor/itest/common/AbstractBaseITests.java
```diff
@@ -418,6 +417,16 @@ public void testControllerView() {
         assertLogEvents(span.logEntries(), Arrays.asList("preHandle", "afterCompletion"));
     }
 
+    @Test
+    public void testExcludePattern() throws InterruptedException {
+        {
+            getRestTemplate().getForEntity("/health", String.class);
+            Thread.sleep(505);
```
jpkrohling: Why this wait?

jpkrohling: And more intriguingly, why `505`?

opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingRestTemplateInterceptorTest.java
```diff
@@ -99,7 +99,7 @@ public void testStandardTagsWithPort() {
         Assert.assertEquals("GET", mockSpan.tags().get(Tags.HTTP_METHOD.getKey()));
         Assert.assertEquals(url, mockSpan.tags().get(Tags.HTTP_URL.getKey()));
         Assert.assertEquals(200, mockSpan.tags().get(Tags.HTTP_STATUS.getKey()));
-        Assert.assertEquals((short)4000, mockSpan.tags().get(Tags.PEER_PORT.getKey()));
```
jpkrohling: This is not related to this change, right ?

opentracing-spring-web-itest/common/src/test/java/io/opentracing/contrib/spring/web/interceptor/itest/common/AbstractBaseITests.java
pavolloffay: I know that you will ask :+1: . Just making sure that span is not created.

> And more intriguingly, why 505? 

It's a nice number but maybe 100ms would be enough.

opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingRestTemplateInterceptorTest.java
pavolloffay: It is, this bumps filter version which uses ot 0.21 where port has been changed to `int`.

jpkrohling: I imagined so, just didn't notice the version bump.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
pavolloffay: `tracingConfiguration` won't be `null` and string concat with null works:
http://stackoverflow.com/questions/4260723/concatenating-null-strings-in-java
https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.11

opentracing-spring-web-itest/common/src/test/java/io/opentracing/contrib/spring/web/interceptor/itest/common/AbstractBaseITests.java
jpkrohling: Isn't there a better way to do that? Like, using Powermock to ensure a method is *not* called?

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebTracingConfiguration.java
pavolloffay: I have changed it to getter. The reason for this is that java is too verbose for this simple props.

pavolloffay:
@jpkrohling thanks!

PR #21, opened 2017-05-29
Bump OT to 0.30.0

pavolloffay:
This should be improved version of https://github.com/opentracing-contrib/java-spring-web/pull/20. Thanks to @objectiser for starting it. 

This PR should better handle activation and deactivation of an active span. It always creates active span in `preHandle` and deactivates in `afterCompletion`.


jpkrohling:
Before I give my OK, would you please confirm there are enough tests in place to assert that the asynchronous aspect works as expected? We discussed some tests for that in the past, but I think it was in the context of other component. The reason I'm asking is because I've seen a few terms that did raise a flag, like the Thread Local (granted, only in tests) and Queue/Dequeue (which I believe is only in the context of a single request, which might then be OK).

pavolloffay:
Note that this code has been using `spanmanager` for active span management. Btw. this is the repo where we discussed that async test (for `AsyncRestTemplate`).

pavolloffay:
@objectiser @jpkrohling thanks!

Issue #23, opened 2017-06-03
check if activeSpanStack is empty before pop


sweatybridge:
The current implementations for TracingHandlerInterceptor.java throws java.util.NoSuchElementException at runtime when the active span stack is empty. The null check happens only after the exception is thrown. The affected lines are:

1. https://github.com/opentracing-contrib/java-spring-web/blob/master/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptor.java#L97
2. https://github.com/opentracing-contrib/java-spring-web/blob/master/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptor.java#L112

sweatybridge:
I believe this is a regression caused by 0.0.5 as the previous version seems to work fine.

pavolloffay:
@sweatybridge do you have a reproducer? It should not happen that stack is empty in `afterConcurrentHandlingStarted` or `afterCompletion`.

sweatybridge:
Yup, I've created a new branch on my benchmark app (https://github.com/sweatybridge/spring-jaeger/tree/bug/stack_pop) that reproduces the issue. You can see the exception after visiting `localhost:8080/a` and inspecting `docker service logs mockdb`.

I've resolved some build problems from previous commits so you should be able to just docker deploy the stack now.

pavolloffay:
@sweatybridge you are using jaeger-client-java of version 0.19.0 which implements OpenTracing java API 0.22.0 and spring instrumentation which implements OT API 0.30.0. 

```
JAEGER_SERVICE_NAME=spring JAEGER_AGENT_HOST=localhost JAEGER_SAMPLER_TYPE=const JAEGER_SAMPLER_PARAM=1 java -jar build/libs/gs-spring-boot-0.1.0.jar

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::        (v1.5.2.RELEASE)

[...]

PR #24, opened 2017-06-05
Remove unnecessary if, stask should not be empty

pavolloffay:
Related to https://github.com/opentracing-contrib/java-spring-web/issues/23. At the moment this if is unnecessary, `pop` would throw an exception.


pavolloffay:
@objectiser thanks

PR #31, opened 2017-08-11
Use spring boot 1.4.1 and consume versions from spring dependencies

pavolloffay:


opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingAsyncRestTemplateTest.java
```diff
@@ -90,6 +90,7 @@ public void run() {
         }
 
         // wait to finish all calls
+        System.out.println(futures.size());
```
jpkrohling: Oops?

pavolloffay: psst nobody saw it!

PR #35, opened 2017-08-20
Implicit constructor injection

eddumelendez:
Spring Framework 4.3 adds some refinements such as implicit constructor
injection and intruduction of ObjectProvider as a programmatic
resolution of dependencies.


pavolloffay:
For a wider compatibility, it's better to don't depend on this feature.

pavolloffay:
@eddumelendez thanks for the PR I will close it, We can reconsider later.

PR #43, opened 2017-10-08
Update README.md

rbtcollins:
Showing the traced span as closed before the servers 'work' is confusing to folk newly encountering opentracing.


pavolloffay:
@rbtcollins thanks

PR #48, opened 2017-12-08
Setting "error" tag in case of exception

elgris:
"error" tag is one of [standard Opentracing tags](https://github.com/opentracing/specification/blob/master/semantic_conventions.md), it indicates that a failure happened during processing of a request. This PR implements setting such a tag in case processing of HTTP request thrown an Exception. Basically, it complements functionality of Client Interceptors, but helps to detect quickly if error happened on server side (useful if the client does not support tracing and the trace originates on the server).

### DIFF
--- a/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/HandlerInterceptorSpanDecorator.java
+++ b/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/HandlerInterceptorSpanDecorator.java
@@ -1,6 +1,7 @@
 package io.opentracing.contrib.spring.web.interceptor;
 
 import io.opentracing.BaseSpan;
+import io.opentracing.tag.Tags;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.method.HandlerMethod;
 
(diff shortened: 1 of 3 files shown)


pavolloffay:
isn't that exception propagated to the servlet filter and logged there? If not I would rather harcoded it in the afterCompletion callback with exception logs.

elgris:
@pavolloffay are you talking about moving this functionality to `onAfterCompletion` method of `HandlerInterceptorSpanDecorator#STANDARD_LOGS` ? I thought about that, just decided to make them separate because `STANDARD_LOGS` is about logs, but `ERROR_TAG` is about tagging.

Does it makes more sense to rename `ERROR_TAG` to `STANDARD_TAGS` and implement processing of some of [standard tags](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table) there along with `error` tag?

pavolloffay:
@elgris all exceptions are logged in servlet filter except exceptions thrown in async callbacks. Those should be also logged via servlet filter https://github.com/opentracing-contrib/java-web-servlet-filter/issues/36

pavolloffay:
I will close for now, feel free to reopen

PR #49, opened 2017-12-10
Make possible to create TracingHandlerInterceptor via spring xml

SerCeMan:
Hi! 

Thank you for the very useful project! I encounter an issue, currently, it is not possible to create an instance of `TracingHandlerInterceptor` via spring xml config. 

From [spring docs](https://docs.spring.io/spring/docs/5.0.x/spring-framework-reference/core.html#beans-autowired-annotation):

> Only one annotated constructor per-class can be marked as required, but multiple non-required constructors can be annotated. In that case, each is considered among the candidates and Spring uses the greediest constructor whose dependencies can be satisfied, that is the constructor that has the largest number of arguments.

`TracingHandlerInterceptor` has 2 constructors annotated with `required=true` (default) which is against the spec. Therefore, during creation the exception will be thrown:
```
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'io.opentracing.contrib.spring.web.interceptor.TracingHandlerInterceptor#0': Invalid autowire-marked constructor: public io.opentracing.contrib.spring.web.interceptor.TracingHandlerInterceptor(io.opentracing.Tracer,java.util.List). Found constructor with 'required' Autowired annotation already: public io.opentracing.contrib.spring.web.interceptor.TracingHandlerInterceptor(io.opentracing.Tracer)
```
The test demostrates the problem.

It can be fixed in 2 ways - annotate both constructors as `required=false` or remove one of them what I've done in this PR.

### DIFF
--- a/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptor.java
+++ b/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptor.java
@@ -48,7 +48,6 @@ public class TracingHandlerInterceptor extends HandlerInterceptorAdapter {
      * @param tracer tracer
      * @param decorators span decorators
      */
-    @Autowired
     public TracingHandlerInterceptor(Tracer tracer, List<HandlerInterceptorSpanDecorator> decorators) {
         this.tracer = tracer;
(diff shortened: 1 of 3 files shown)


SerCeMan:
Hey, @pavolloffay! Can you take a look at this PR?

pavolloffay:
@SerCeMan thanks!

PR #50, opened 2018-01-11
Remove unnecessary @Autowired annotation from constructor

geoand:
Since TracingHandlerInterceptor is not annotated with one of the Spring
stereotype annotations but is instead meant to be created using explicit
Java or XML Configuration, the @Autowired annotation on the constructor
is no longer needed

### DIFF
--- a/opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptorJavaConfigIntegrationTest.java
+++ b/opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptorJavaConfigIntegrationTest.java
@@ -1,0 +1,41 @@
+package io.opentracing.contrib.spring.web.interceptor;
+
+import io.opentracing.Tracer;
+import io.opentracing.mock.MockTracer;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptorJavaConfigIntegrationTest.java
```diff
@@ -0,0 +1,41 @@
...
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = TracingHandlerInterceptorJavaConfigIntegrationTest.TestConfiguration.class)
+public class TracingHandlerInterceptorJavaConfigIntegrationTest {
+
+    @Autowired
+    private TracingHandlerInterceptor interceptor;
+
+    @Test
+    public void testAutowired() {
+        assertNotNull(interceptor);
```
pavolloffay: What does this test actually do?

geoand: The same thing that the existing XML configuration test does. It ensures that the `TracingHandlerInterceptor` can be created correctly when using Java Config (as the XML test did for the XML configuration)

pavolloffay:
@geoand thanks I forgot about this PR

geoand:
@pavolloffay Me too :P

PR #62, opened 2018-02-20
Improve configurability of the ServerTracingAutoConfiguration

ledor473:
# Changes
While the current default were good, we needed more configuration for some project:
1. `order` and `urlPatterns` are now configurable
2. Project-specific ServletFilterSpanDecorator / HandlerInterceptorSpanDecorator beans can be provided to the ServerTracingAutoConfiguration (via ObjectProvider [1])

# Why this change?
1. `urlPatterns` can be useful to trace only some specific part of an application. As for `order`, it's always useful to have it in case you want to have another filter before this one.
2. While the SpanDecorator pattern is very nice, there was no way to use the ServerTracingAutoConfiguration along with a custom SpanDecorator.  

# Reference
[1] The usage of ObjectProvider is similar to the [RestTemplateAutoConfiguration of Spring Boot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/RestTemplateAutoConfiguration.java)

cc @jfbreault

### DIFF
--- a/opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
+++ b/opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
@@ -25,6 +25,8 @@ import io.opentracing.contrib.spring.web.interceptor.TracingHandlerInterceptor;
 import io.opentracing.contrib.web.servlet.filter.ServletFilterSpanDecorator;
 import io.opentracing.contrib.web.servlet.filter.TracingFilter;
 
+import static java.lang.String.format;
+
 /**
  * @author Pavol Loffay
  * @author Eddú Meléndez
@@ -49,9 +51,10 @@ public class ServerTracingAutoConfiguration {
     @Bean
     @ConditionalOnMissingBean(TracingFilter.class)
     public FilterRegistrationBean tracingFilter(Tracer tracer, WebTracingProperties tracingConfiguration) {
-        log.info("Creating " + FilterRegistrationBean.class.getSimpleName() + " bean with " +
-                TracingFilter.class + " mapped to " + tracingConfiguration.getUrlPatterns().toString() +
-                ", skip pattern is " + tracingConfiguration.getSkipPattern());
+        log.info(format("Creating %s bean with %s mapped to %s, skip pattern is \"%s\"",
+                FilterRegistrationBean.class.getSimpleName(), TracingFilter.class.getSimpleName(),
+                tracingConfiguration.getUrlPatterns().toString(),
+                tracingConfiguration.getSkipPattern()));
 
         List<ServletFilterSpanDecorator> decorators = servletFilterSpanDecorator.getIfAvailable();
         if (CollectionUtils.isEmpty(decorators)) {
--- a/opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
+++ b/opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
@@ -2,9 +2,11 @@ package io.opentracing.contrib.spring.web.autoconfig;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -13,6 +15,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.util.CollectionUtils;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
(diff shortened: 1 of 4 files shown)


opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
```diff
@@ -34,18 +37,32 @@
 public class ServerTracingAutoConfiguration {
     private static final Log log = LogFactory.getLog(ServerTracingAutoConfiguration.class);
 
+    private final ObjectProvider<List<ServletFilterSpanDecorator>> servletFilterSpanDecorator;
```
pavolloffay: `ObjectProvider` is since 4.3  https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/ObjectProvider.html. Is there a way to do it without it? We want to support the largest range of spring versions if possible.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
```diff
@@ -34,18 +37,32 @@
...
+    public ServerTracingAutoConfiguration(ObjectProvider<List<ServletFilterSpanDecorator>> servletFilterSpanDecorator,
+                                          ObjectProvider<List<HandlerInterceptorSpanDecorator>> interceptorSpanDecorator) {
+        this.servletFilterSpanDecorator = servletFilterSpanDecorator;
+        this.interceptorSpanDecorator = interceptorSpanDecorator;
+    }
+
     @Bean
     @ConditionalOnMissingBean(TracingFilter.class)
     public FilterRegistrationBean tracingFilter(Tracer tracer, WebTracingProperties tracingConfiguration) {
         log.info("Creating " + FilterRegistrationBean.class.getSimpleName() + " bean with " +
```
jfbreault: Since not using slf4j, we should wrap that in `log.isInfoEnabled()` since we are processing some stuff to log

pavolloffay:
> While the SpanDecorator pattern is very nice, there was no way to use the ServerTracingAutoConfiguration along with a custom SpanDecorator. 

I was thinking about deprecating `SpanDecorators` as the same can be achieved with standard filters/interceptors.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
ledor473: Spring 4.3 [is part of Spring Boot 1.4.x](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-1.4-Release-Notes#spring-framework-43) which is the version specified in the [pom.xml](https://github.com/opentracing-contrib/java-spring-web/blob/master/pom.xml#L48)

pavolloffay: thanks, I was more talking about this https://github.com/opentracing-contrib/java-spring-web/tree/master/opentracing-spring-web package. I forgot that this is auto-configure.

ledor473:
As for deprecating `SpanDecorators`, I feel like it's more work to implement it in a Filter correctly (async vs sync) than a Decorator. With a custom Filter, the order will also be very important.  
It would probably result in more code and config duplication (async handling, accessing the span from the request's attribute, same excludePatterns and urlPatterns for both Filters).

Internally, we've used the same SpanDecorator pattern for JMS and AOP (both JDBC and `@Traced`) and were looking to contribute that upstream.

Here's two Decorator that we are using:
- HeaderSpanDecorator: get a list of Header and make it a span tag
- MDCSpanDecorator: get a list of MDC entry and make it a span tag

jfbreault:
Not JMS @ledor473 but RestTemplate. Hopefully JMS and Jdbc could use the same pattern since as of now, we cannot cutomize them

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
pavolloffay: I guess that is fine as this is invoked only at app start time.

`String.format() ` should help with readability.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebTracingProperties.java
```diff
@@ -22,4 +28,21 @@ public Pattern getSkipPattern() {
     public void setSkipPattern(Pattern skipPattern) {
         this.skipPattern = skipPattern;
     }
+
+    public List<String> getUrlPatterns() {
+        return urlPatterns;
+    }
+
+    public void setUrlPatterns(List<String> urlPatterns) {
```
pavolloffay: Can this be specified in spring config e.g. yml?

jfbreault: Yes, that's the main point.
```
opentracing:
  spring:
    web:
      url-patterns:
        - /*
        - ...
```

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
jfbreault: Maybe but a bad practice especialy in lib, Spring wrap error in isErrorEnabled() if message need to be processing!

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebTracingProperties.java
ledor473: @pavolloffay the test is in place to ensure the list is loaded correctly (I added application-test.yml)

ledor473:
Updated the PR with `String.format()`

pavolloffay:
@ledor473 @jfbreault thanks!

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfigurationTest.java
```diff
@@ -42,14 +50,48 @@ public MockTracer tracer() {
...
     private MockTracer mockTracer;
 
+    @Autowired
+    @Qualifier("tracingFilter")
+    private FilterRegistrationBean tracingFilter;
+
+    @MockBean
+    private ServletFilterSpanDecorator mockDecorator1;
+    @MockBean
+    private ServletFilterSpanDecorator mockDecorator2;
```
pavolloffay: @ledor473 @jfbreault can you explain why there are these two decorators? 

Is it the decorator used by the filter? Then there is only one.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebTracingProperties.java
pavolloffay: @jfbreault I could make it work only like:

`urlPatterns: /a, /b`

https://github.com/opentracing-contrib/java-spring-web/pull/63/files#diff-2321c00ebd91c85cf1819d2606f08147R4

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfigurationTest.java
ledor473: There's 2 Decorators to ensure that the ObjectProvider can correctly provide more than 1 Decorator to the Filter. 

It's used by this test https://github.com/opentracing-contrib/java-spring-web/pull/62/files/09692c02b88b35acb4801ffe7e7ea86a8f49164d#diff-86e82c5211f5ddaf9510408e7ea71310R86

pavolloffay: Oh I see it's providing a bean for the auto-configuration. sorry

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebTracingProperties.java
jfbreault: My exemple seem wrong, the `-` need to be on the same indent as the parent so something like this:
```
opentracing:
  spring:
    web:
      url-patterns:
      - /*
      - ...
```

PR #64, opened 2018-03-01
Fix to correctly use a List<> with Spring @Properties

ledor473:
@pavolloffay I was wondering why you had to change the `application-test.yml` in https://github.com/opentracing-contrib/java-spring-web/pull/63 to make it work.

I guess it's because it was throwing a `java.lang.UnsupportedOperationException` when starting the context?

The problem was that when an List is already initialized in the Properties object, Spring add the items to it instead of replacing it by a new List. 
But the issue was that I used `Arrays.asList()` which [returns a fixed-size list](https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList%28T...%29)


pavolloffay:
@ledor473 thanks

PR #66, opened 2018-04-04
Feature/65 disable web tracing

ask4gilles:


opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
```diff
@@ -35,7 +34,6 @@
 @ConditionalOnWebApplication
 @ConditionalOnBean(Tracer.class)
 @AutoConfigureAfter(TracerAutoConfiguration.class)
-@EnableConfigurationProperties(WebTracingProperties.class)
```
pavolloffay: Only this auto-configuration should be affected by `opentracing.spring.web.enabled`

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/TracerAutoConfiguration.java
```diff
@@ -15,6 +17,8 @@
  * @author Pavol Loffay
  */
 @Configuration
+@EnableConfigurationProperties(WebTracingProperties.class)
+@ConditionalOnProperty(name = "opentracing.spring.web.enabled", havingValue = "true", matchIfMissing = true)
```
pavolloffay: I don't think it's needed here. This configuration provides tracer bean for different instrumentations and auto-configs

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
ask4gilles: Hi @pavolloffay,

The problem is that if I just disable the configuration in ServerTracingAutoConfiguration and not in the TracerAutoConfiguration, the AsyncRestTemplateAutoConfiguration and RestTemplateAutoConfiguration will still kick in and you will have RestTemplates instrumented with a NoopTracer, which according to me, does not make any sense: if you want to disable web tracing, why would you want your rest templates to be instrumented?

The alternate solution would imply cluttering these rest templates auto configurations with:
```
@EnableConfigurationProperties(WebTracingProperties.class)
@ConditionalOnProperty(name = "opentracing.spring.web.enabled", havingValue = "true", matchIfMissing = true)
``` ?

pavolloffay: Adding it to `TracerAutoConfiguration` is just wrong it breaks other stuff in OT spring cloud. 

I would argue that tracing in web clients should have a separate config property. (right now one can disable tracing in web client by not auto-wiring it and just creating a direct instance)

ask4gilles:
@pavolloffay Alright, I pushed the desired modifications. Cheers.

PR #67, opened 2018-04-05
Generate spring-configuration-metadata.json for code autocompletion

mdvorak:
Added spring-boot-configuration-processor as Maven dependency
When metadata is missing, IDEA reports correct configuration as warning


pavolloffay:
@mdvorak thanks! We should do the same in ot-spring-cloud

PR #68, opened 2018-04-06
Configurable RestTemplate tracing with support for RestTemplateBuilder

mdvorak:
1. Added `WebClientTracingProperties`, with `enabled` and `componentName`
2. Moved `RestTemplate` and `AsyncRestTemplate` AutoConfig into once umbrella class `RestTemplateTracingAutoConfiguration`, which allows us to have all `@Conditional` annotations in one place. Actual configurations are in nested static classes, which is pattern used heavily by Spring Boot itself. Allows us to use additional conditions on each configuration class.
3. Added `TracingRestTemplateCustomizer`, which is used by Spring Boot `RestTemplateBuilder` (not included in Spring Framework itself). With SB, its valid pattern to have one preconfigured builder bean, and consumers create `RestTemplate` instances locally, private, even on demand - but still share common configuration. `RestTemplateCustomizer` is pattern to hook into builder config without directly modifying the bean. Side effect is that also `TestRestTemplate` now creates traces - therefore disabled client tracing for Servlet tests.

This is a big PR including multiple changes, but I found it would be complicated to split it across several PRs. 
It should however be non-breaking backwards compatible. Only change in behavior should be for `RestTemplateBuilder` beans, which might influence some applications. But its likely it will fix them instead of breaking them.

Discussion is welcome :)


pavolloffay:
@mdvorak thanks for the PR.

The auto-configuration part looks fine. 

* can we keep the core module without spring boot classes e.g. `RestTemplateBuilder`? 
* Could you keep the instrumentation classes without changes?

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebClientTracingProperties.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * When set to true (default), it enables automatic tracing of RestTemplate beans, as well as instances created using default RestTemplateBuilder bean.
+     * Does not affect instances created manually.
+     */
+    private boolean enabled = true;
+    /**
+     * Configures COMPONENT tag value, when default RestTemplateSpanDecorator.StandardTags bean is used.
+     * When custom RestTemplateSpanDecorator is provided, this configuration has no effect.
+     */
+    private String componentName = RestTemplateSpanDecorator.StandardTags.COMPONENT_NAME;
```
pavolloffay: Do you need this to be configurable?

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/DisabledRestTemplateTracingAutoConfigurationTest.java
```diff
@@ -0,0 +1,104 @@
...
+    @Autowired(required = false)
+    private FilterRegistrationBean tracingFilter;
+    @Autowired(required = false)
+    private WebMvcConfigurerAdapter tracingHandlerInterceptor;
+    @Autowired
+    private MockTracer mockTracer;
+    @Autowired
+    private RestTemplate restTemplate;
+    @Autowired
+    private AsyncRestTemplate asyncRestTemplate;
```
pavolloffay: It could also test that there is no tracing interceptor installed?

pavolloffay: if it is possible

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/SkipPatternTest.java
```diff
@@ -14,21 +10,26 @@
...
 /**
  * @author Pavol Loffay
  */
 @SpringBootTest(
         webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
         classes = {SkipPatternTest.SpringConfiguration.class, SkipPatternTest.Controller.class},
-        properties = {"opentracing.spring.web.skipPattern=/skip"})
+        properties = {
+                "opentracing.spring.web.skipPattern=/skip",
+                "opentracing.spring.web.client.enabled=false"
```
pavolloffay: Is this necessary? if not remove it

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/RestTemplateSpanDecorator.java
```diff
@@ -60,11 +61,21 @@
     class StandardTags implements RestTemplateSpanDecorator {
         private static final Log log = LogFactory.getLog(StandardTags.class);
 
-        public static String COMPONENT_NAME = "java-spring-rest-template";
+        public static final String COMPONENT_NAME = "java-spring-rest-template";
+
+        private final String componentName;
+
+        public StandardTags() {
+            this(null);
```
pavolloffay: Passing `COMPONENT_NAME` makes it cleaner, at least for me

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/TracingAsyncRestTemplateInterceptor.java
```diff
@@ -14,14 +15,14 @@
 import org.springframework.util.concurrent.ListenableFuture;
 import org.springframework.util.concurrent.ListenableFutureCallback;
 
-import io.opentracing.Scope;
-import io.opentracing.Span;
-import io.opentracing.Tracer;
-import io.opentracing.propagation.Format;
-import io.opentracing.tag.Tags;
-import io.opentracing.util.GlobalTracer;
+import java.io.IOException;
```
pavolloffay: there shouldn't be changes in this file

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/RestTemplateSpanDecorator.java
mdvorak: And I was looking at this line for at least a minute :)

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/DisabledRestTemplateTracingAutoConfigurationTest.java
mdvorak: Added in 05446e26d38bce5cca62092be88c09adb1efa6ad

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebClientTracingProperties.java
pavolloffay: I don't have any objections to make it configurable, I am just wondering why do you want to override it. Do yo have any specific use case?

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/SkipPatternTest.java
pavolloffay: If it is necessary then the behaviour has changes I guess

pavolloffay:
Can be `TracingRestTemplateCustomizer`  moved to the starter module?

mdvorak:
You mean `opentracing-spring-web-autoconfigure`? Of course.
Moving it to `opentracing-spring-cloud` would split RestTemplate config into two separate projects, which I don't like.
Commit comming up :)

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/SkipPatternTest.java
mdvorak: Yes, it modifies behavior of everything that comes out of `RestTemplateBuilder` provided by Spring Boot. At adds tracing to those instances as well.

Interally, `TestRestTemplate` uses by default this pattern, with private `RestTemplate`.
So, it broke this test, because `TestRestTemplate` produced traces on its own, while test expected exactly one trace from the server.
Another way would be to filter out logged spans, but this was easier and more readable IMO.

This change to `TestRestTemplate` should not break any tests in actual applications.

mdvorak:
Note: While `RestTemplateCustomizer` *should* not break anything, its a new feature, and should at least have version bump to `0.3.x`.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/WebClientTracingProperties.java
mdvorak: I think after more digging into spec that you are correct, there is no real use case to modify this name.
Other libraries have it hard-coded as well. I'll take that as convention and drop it.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/RestTemplateTracingAutoConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+            return new RestTemplateSpanDecorator.StandardTags();
+        }
+    }
+
+    /**
+     * Injects {@link TracingRestTemplateInterceptor} into {@link InterceptingHttpAccessor#getInterceptors()}.
+     */
+    @Configuration
+    @ConditionalOnBean(InterceptingHttpAccessor.class)
+    public static class RestTemplatePostProcessingConfiguration {
```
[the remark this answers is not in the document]
mdvorak: Its actual base class that provides interceptor support for RestTemplate. Its questionable, whether limit it to RestTemplate or use this potentionally wider base class.

Conditional on Async... is because its deprecated and will be removed probably in Spring 6.x, therefore ConditionalOnClass makes it forward compatible.

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/CustomSpanDecoratorAutoConfigurationTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Autowired
+    private AsyncRestTemplate asyncRestTemplate;
+
+    @Before
+    public void setUp() {
+        mockTracer.reset();
+    }
+
+    @Test
+    public void testRestClientNotTracing() {
```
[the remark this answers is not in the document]
mdvorak: Now that is copy-paste error, I'll rename the methods.

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/SkipPatternTest.java
pavolloffay: Let's keep this, sleuth also instruments test template

mdvorak: Well, that is possible, since you are actually testing spans there as well. Still not typical use case outside OT tests thou.
It is technically correct behavior, since you take TestRestTemplate from Spring context that has tracing enabled, so its logical it traces the requests.. But it is common practice for Spring Boot apps.
I'll try to run tests with snapshot, to see what happens. I'm afraid that it won't be sooner than Monday.

mdvorak:
I've tested the change with opentracing-spring-cloud, and indeed there are plenty of failing tests - whenever they expect exact span count in the tracer and TestRestTemplate is used.
I can prepare patch for *otsc* that will work with this version by disabling client tracing for all tests where servlet is tested.

+ due to this change, i would really bump version to 0.3.0, what do you thing @pavolloffay? I can change the version in PR if you want.

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/RestTemplateTracingAutoConfiguration.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Provides bean {@link RestTemplateSpanDecorator.StandardTags}.
+     */
+    @Configuration
+    @ConditionalOnMissingBean(RestTemplateSpanDecorator.class)
+    public static class StandardTagsConfiguration {
+
+        @Bean
+        public RestTemplateSpanDecorator.StandardTags standardTagsRestTemplateSpanDecorator() {
```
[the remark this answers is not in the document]
mdvorak: You have same logic available (with a bit different syntax) for `ServletFilterSpanDecorator` and `HandlerInterceptorSpanDecorator`. Since `RestTemplateSpanDecorator` is an interface with one standard implementation, it looks like it should be overridable.
And while I don't currently will be overriding this behavior, I have in mind case where I might need it (as a bridge to legacy system).

pavolloffay:
@mdvorak sorry for the delay, I am fine with changes, I will have the last look and merge/comment. 

You don't have to create a PR to bump the version, it will be done automatically as part of the release.

mdvorak:
No problem, thanks for the review, I know its as hard as coding :-)

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/RestTemplateTracingAutoConfiguration.java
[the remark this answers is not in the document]
pavolloffay: I guess the behavior to server decorators is a bit different. On the server-side user can disable standard tags whereas here standard tags will be always there.

pavolloffay:
@mdvorak thanks! I will cut 0.3.0

opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/RestTemplateTracingAutoConfiguration.java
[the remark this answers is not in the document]
mdvorak: `@ConditionalOnMissingBean(RestTemplateSpanDecorator.class)` on Configuration class makes this config fallback whenever another bean is provided. So behavior is same in the and.
Originally there were more code, that's why it is on Config class instead on Bean. But results is same.

opentracing-spring-web-autoconfigure/src/test/java/io/opentracing/contrib/spring/web/autoconfig/SkipPatternTest.java
mdvorak: Just to close this, I've run tests with ot-spring-cloud and it indeed will need changes. Adding this property to those tests as well will fix it.

PR #70, opened 2018-04-24
Increase timeout of test because it sometimes fails on PNC

geoand:
@pavolloffay @objectiser Could you merge this and then release a new version please so I can use that in PNC?

Thanks


objectiser:
@geoand Version 0.3.2 on its way.

geoand:
Wonderful, thanks a lot!

PR #72, opened 2018-05-29
Fix runtime exception when running with webflux

### LINKED ISSUE #59: Fails when using Spring Webflux
Are you aware this fails with Spring Webflux?

pavolloffay:
Resolves #59 

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

### DIFF
--- a/opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
+++ b/opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
@@ -7,6 +7,7 @@ import java.util.List;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
@@ -28,6 +29,8 @@ import org.apache.commons.logging.LogFactory;
 
 import static java.lang.String.format;
 
+import javax.servlet.Servlet;
+
 /**
  * @author Pavol Loffay
  * @author Eddú Meléndez
@@ -37,6 +40,7 @@ import static java.lang.String.format;
 @ConditionalOnBean(Tracer.class)
 @AutoConfigureAfter(TracerAutoConfiguration.class)
 @EnableConfigurationProperties(WebTracingProperties.class)
+@ConditionalOnClass(Servlet.class)
 @ConditionalOnProperty(name = "opentracing.spring.web.enabled", havingValue = "true", matchIfMissing = true)
 public class ServerTracingAutoConfiguration {
     private static final Log log = LogFactory.getLog(ServerTracingAutoConfiguration.class);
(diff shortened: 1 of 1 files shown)


opentracing-spring-web-autoconfigure/src/main/java/io/opentracing/contrib/spring/web/autoconfig/ServerTracingAutoConfiguration.java
```diff
@@ -37,6 +40,7 @@
 @ConditionalOnBean(Tracer.class)
 @AutoConfigureAfter(TracerAutoConfiguration.class)
 @EnableConfigurationProperties(WebTracingProperties.class)
+@ConditionalOnClass(Servlet.class)
```
pavolloffay: IN SB 2 it's possible to write 

`@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)`

pavolloffay:
PS I have tested it locally with hello world style webflux app. It didn't crash

pavolloffay:
@darrenhaken could you please verify this?

darrenhaken:
@pavolloffay sure thing, what version should I pull down to test against?

darrenhaken:
@pavolloffay the runtime exception doesn't occur with a WebFlux API.

We did experience an issue which I thought you should be aware of; it's an edge case though based on our environment. In order to support a legacy component in our platform, we had to add into the servlet API dependency to the WebFlux based service.

I thought it was worth mentioning because it's hard to say how much of an edge case this is.

This did start to cause issues again. Not sure if there is a more elegant code change which would prevent this from happening. Exception as follows:

```
Caused by: java.lang.NoClassDefFoundError: org/springframework/web/servlet/config/annotation/WebMvcConfigurerAdapter
    at java.lang.Class.getDeclaredMethods0(Native Method) ~[na:1.8.0_144]
    at java.lang.Class.privateGetDeclaredMethods(Class.java:2701) ~[na:1.8.0_144]
    at java.lang.Class.getDeclaredMethods(Class.java:1975) ~[na:1.8.0_144]
    at org.springframework.util.ReflectionUtils.getDeclaredMethods(ReflectionUtils.java:641) ~[spring-core-5.0.6.RELEASE.jar:5.0.6.RELEASE]
    ... 25 common frames omitted
Caused by: java.lang.ClassNotFoundException: org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter
    at java.net.URLClassLoader.findClass(URLClassLoader.java:381) ~[na:1.8.0_144]
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424) ~[na:1.8.0_144]
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335) ~[na:1.8.0_144]
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357) ~[na:1.8.0_144]
    ... 29 common frames omitted
```

pavolloffay:
@darrenhaken thanks!

If you have it locally cloned, would you like to try using `@ConditionalOnClass(WebMvcConfigurerAdapter.class)` instead of `Servlet.class`?

darrenhaken:
@pavolloffay Will check that today

timpharo:
@pavolloffay following on from discussions with @darrenhaken i have retested this locally with `@ConditionalOnClass(WebMvcConfigurerAdapter.class)` and this fixes our issues regarding startup once more, cheers.  

That said i am still not seeing any tracing hitting Jaeger unless i explicitly declare a span within one of the routes.

pavolloffay:
@timpharo thanks for the test. 

> That said i am still not seeing any tracing hitting Jaeger unless i explicitly declare a span within one of the routes.

That is expected, webflux will need a different instrumentation.

darrenhaken:
What's the different instrumentation needed for Webflux?

geoand:
Webflux is an entirely different web stack (a reactive one based based on
project reactor), it has nothing to do with Web MVC (which is servlet
based), therefore it needs an entirely new instrumentation.

darrenhaken:
Would you imagine this library to add instrumentation or is it lower down the stack like the reactor project?

geoand:
I would imagine that instrumentation will be added at some point in time
and it would probably be based on some of the context propagation features
in Reactor (but that is just a guess 😀)

pavolloffay:
Here is an issue to track it  https://github.com/opentracing-contrib/java-spring-cloud/issues/150.

https://docs.spring.io/spring/docs/5.0.0.BUILD-SNAPSHOT/spring-framework-reference/html/web-reactive.html#web-reactive-server it mentions that webflux can run on servlet containers. If it's your case you can servlet filter integration https://github.com/opentracing-contrib/java-web-servlet-filter, or use the filter from this repository.

PR #79, opened 2018-08-30
WIP: Fix client span finishes before server

### LINKED ISSUE #78: Client span is finished before server span
screenshot of jaeger ui

`GET` is a client request. `get` is a  server span for that request. The server span should be finished before the client request.

cc @yurishkuro

pavolloffay:
Resolves #78 

Do not merge yet! The issue is still present

Signed-off-by: Pavol Loffay <ploffay@redhat.com>


wuyupengwoaini:
I find the same problem and I want to know when to merge and what is the real reason to cause this problem though I view the source and then I can not understand .Could you please tell me the real reason

PR #81, opened 2018-09-30
Fix premature closing span

### LINKED ISSUE #80: TracingHandlerInterceptor closes the Scope, even if it hasn't been opened in preHandle()
I'm using a spring boot application that has several filters. 

I observe the following behavior: going in with the request the scope was set by TracingFilter, processing the response the scope was already closed. Looking at TracingHandlerInterceptor it looks like it closes the scope too early after processing the request

----

Debuggig the situation I found that TracingHandlerInterceptor *only sometimes* opens a scope in preHandle()

```
Scope serverSpan = tracer.scopeManager().active();
if (serverSpan == null) {
        /* ... open scope ... */ 
}
```

but it will *always* close the scope in afterCompletion()

```
Deque<Scope> scopeStack = getScopeStack(httpServletRequest);
Scope scope = scopeStack.pop();
/* ... */
scope.close();
```

When you put a breakpoint in ThreadLocalScope you'll see that the following condition will be true when TracingFilter tries to close the scope again.

```
if (scopeManager.getTlsScope() != this) {
    // This shouldn't happen if users call methods in the expected order. Bail out.
    return;
}
```
----

I'm using opentracing-spring-jaeger-cloud-starter:0.2.1 that includes opentracing-web-servlet-filter:0.1.0 and opentracing-spring-web:0.3.3. This also happens with opentracing-web-servlet-filter:0.2.0

ahus1:
I've added a SpanCheckerFilter to check the condition that a span shouldn't be closed prematurely. Then I modified TracingHandlerInterceptor to make the tests pass.

@pavolloffay - Let me know what you think.

closes #80


geoand:
@ahus1 Thanks for the PR!

I'll defer to @pavolloffay for reviewing since he's the expert here, but could I suggest that you add an integration test that shows in code the problem stated in #80 and that this fix actually addresses it?

geoand:
@ahus1 Never mind my comment about the test, I just saw that it you added one :P

opentracing-spring-web-itest/mvc/src/test/webapp/WEB-INF/web.xml
```diff
@@ -52,6 +52,17 @@
     <url-pattern>/*</url-pattern>
   </filter-mapping>
 
+  <!-- exception filter -->
```
pavolloffay: It seems like copied from the above

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/TracingHandlerInterceptor.java
```diff
@@ -1,25 +1,16 @@
 package io.opentracing.contrib.spring.web.interceptor;
 
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Deque;
-import java.util.List;
+import io.opentracing.*;
```
pavolloffay: avoid wildcard imports

ahus1:
I've updated the code and pushed an updated commit: 

* the web.xml comment has been updated
* the TracingHandlerInterceptor doesn't contain wildcard imports any more

Good to go?

geoand:
@pavolloffay Should we do a release with this fix?
I would like to do a release of java-spring-jaeger soon and it would be good to include all the goodies 😁

pavolloffay:
+1 we should also bump other dependencies in this repo e.g. servlet filter

geoand:
Thanks @pavolloffay! I will do that this week then

PR #82, opened 2018-10-06
An example in README.md is not correct

peterjurkovic:
Hi there, there is no such a class like `ActiveSpan` it should be `Span`. Instead of `deactivate` should be `finish`


geoand:
Thanks @peterjurkovic !

PR #95, opened 2019-02-22
Add Spring WebFlux and WebClient tracing.

csabakos:
This adds Spring WebFlux and WebClient tracing.

I will also add auto-configuration for https://github.com/opentracing-contrib/java-reactor to `opentracing-spring-cloud`, which will complement these changes with automatic activation of the span in handler functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 # OpenTracing Spring Web Instrumentation
 
-This library provides instrumentation for Spring  Web applications (Boot and MVC). It creates tracing data for 
-server requests and also client requests (`RestTemplate` and `AsyncRestTemplate`).
+This library provides instrumentation for Spring Web applications (Boot, MVC and WebFlux). It creates tracing data for 
+server requests and also client requests (`RestTemplate`, `AsyncRestTemplate` and `WebClient`).
 
 ## Use [opentracing-spring-cloud](https://github.com/opentracing-contrib/java-spring-cloud) instead
 
@@ -11,12 +11,22 @@ As it was mentioned above this library traces only inbound/outbound HTTP request
 get automatically traced different set of technologies e.g. `spring-cloud-netflix`, JMS or even more then
 use project [opentracing-spring-cloud](https://github.com/opentracing-contrib/java-spring-cloud) instead.
 
+For reactive applications, it is especially recommended to use `reactor` tracing from
+[opentracing-spring-cloud](https://github.com/opentracing-contrib/java-spring-cloud), as that will ensure
+that the `Span` is activated in reactor handler functions. (Without that, one would have to extract the
+`Span` from the subscriber context.)
+
 ## How does the server tracing work?
 
+### Servlet
 Server span is started in [Web Servlet Filter](https://github.com/opentracing-contrib/java-web-servlet-filter),
 then tracing interceptor adds spring related tags and logs. There are use case when spring boot invokes a handler after 
 a request processing in filter finished, in this case interceptor starts a new span as `followsFrom` 
 which references the initial span created in the servlet filter.
+
+### Reactive
+Server span is started in [TracingWebFilter](opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingWebFilter.java)
+(upon subscription), then `onNext()`, `onError()`, etc. handlers add Spring WebFlux related tags and logs.
 
 ## Library versions
 
@@ -36,11 +46,11 @@ If you are using Spring Boot the easiest way how to configure OpenTracing instru
 
 ```
 Just provide an OpenTracing tracer bean and all required configuration is automatically
-done for you. It also instruments all `RestTemplate` and `AsyncRestTemplate` beans.
+done for you. It also instruments all `RestTemplate`, `AsyncRestTemplate`, `WebClient` and `WebClient.Builder` beans.
 
 ### Manual configuration
 
-#### Server
+#### Servlet and MVC Server
 Configuration needs to add `TracingFilter` and `TracingHandlerInterceptor`. Both of these classes
 are required!
 
@@ -77,12 +87,38 @@ public class MVCConfiguration extends WebMvcConfigurerAdapter {
 
 XML based configuration can be used too. Filter can be also directly defined in `web.xml`.
 
+#### Reactive Server
+Configuration needs to add the `TracingWebFilter` bean.
+
+```java
+@Configuration
+class TracingConfiguration {
+    @Bean
+    public TracingWebFilter tracingWebFilter(Tracer tracer) {
+        return new TracingWebFilter(
+                tracer,
+                Integer.MIN_VALUE,               // Order
+                Pattern.compile(""),             // Skip pattern
+                Collections.singletonList("/*"), // URL patterns
+                Arrays.asList(new WebFluxSpanDecorator.StandardTags(), new WebFluxSpanDecorator.WebFluxTags())
+        );
+    }
+}
+```
+
 #### Client
 ```java
 RestTemplate restTemplate = new RestTemplate();
 restTemplate.setInterceptors(Collections.singletonList(new TracingRestTemplateInterceptor(tracer)));
 
 // the same applies for AsyncRestTemplate 
+```
+
+#### Reactive Client
+```java
+WebClient webClient = WebClient.builder()
+        .filter(new TracingExchangeFilterFunction(tracer, Collections.singletonList(new WebClientSpanDecorator.StandardTags())))
+        .build();
 ```
 
 ## Access server span
--- a/opentracing-spring-web-itest/webflux/src/test/java/io/opentracing/contrib/spring/webflux/webfilter/itest/webflux/WebFluxJettyMinimalIntegrationTest.java
+++ b/opentracing-spring-web-itest/webflux/src/test/java/io/opentracing/contrib/spring/webflux/webfilter/itest/webflux/WebFluxJettyMinimalIntegrationTest.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2016-2018 The OpenTracing Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.opentracing.contrib.spring.webflux.webfilter.itest.webflux;
+
+import io.opentracing.contrib.spring.web.webfilter.TracingWebFilter;
+import io.opentracing.contrib.spring.web.webfilter.WebFluxSpanDecorator;
+import io.opentracing.mock.MockSpan;
+import io.opentracing.mock.MockTracer;
+import io.opentracing.tag.Tags;
+import io.opentracing.util.ThreadLocalScopeManager;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.springframework.boot.test.web.client.TestRestTemplate;
+import org.springframework.boot.web.client.RestTemplateBuilder;
+import org.springframework.boot.web.embedded.jetty.JettyReactiveWebServerFactory;
+import org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext;
+import org.springframework.core.io.buffer.DefaultDataBufferFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.server.reactive.HttpHandler;
+import org.springframework.http.server.reactive.ServerHttpResponse;
+import org.springframework.web.server.ServerWebExchange;
+import org.springframework.web.server.WebHandler;
+import org.springframework.web.server.adapter.WebHttpHandlerBuilder;
+import reactor.core.publisher.Mono;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * @author Csaba Kos
+ */
+public class WebFluxJettyMinimalIntegrationTest {
+
+    private static final ReactiveWebServerApplicationContext APPLICATION_CONTEXT = new ReactiveWebServerApplicationContext();
+    private static final MockTracer MOCK_TRACER = Mockito.spy(new MockTracer(new ThreadLocalScopeManager(),
+            MockTracer.Propagator.TEXT_MAP));
+    private final static TracingWebFilter TRACING_WEB_FILTER = new TracingWebFilter(
+            MOCK_TRACER,
+            Integer.MIN_VALUE,
+            Pattern.compile(""),
+            Collections.singletonList("/*"),
+            Arrays.asList(new WebFluxSpanDecorator.StandardTags(), new WebFluxSpanDecorator.WebFluxTags())
+    );
+    private static final int SERVER_PORT;
+    private static final TestRestTemplate TEST_REST_TEMPLATE;
+
+    static {
+        APPLICATION_CONTEXT.registerBean("jettyReactiveWebServerFactory", JettyReactiveWebServerFactory.class, () ->
+                new JettyReactiveWebServerFactory(0));
+        APPLICATION_CONTEXT.registerBean("webHandler", WebHandler.class, () ->
+                serverWebExchange -> TRACING_WEB_FILTER.filter(serverWebExchange, WebFluxJettyMinimalIntegrationTest::handler));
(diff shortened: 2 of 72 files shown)


opentracing-spring-web/pom.xml
```diff
@@ -37,5 +42,17 @@
       <version>${version.org.springframework}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock-jre8</artifactId>
```
csabakos: I had to replace `MockRestServiceServer` in the tests with something else (WireMock), because `MockRestServiceServer` does not support `WebClient`.

pom.xml
```diff
@@ -35,8 +35,8 @@
   </developers>
 
   <properties>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
```
csabakos: Spring Framework 5 requires JRE 8 so I believe this is safe to change.

pavolloffay: Note that this is not only SB, the web package can be used with old spring web apps.

geoand: @pavolloffay Spring Framework 5 + required JDK 8+ see [here](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-5.x#upgrading-to-version-50)

README.md
```diff
@@ -2,22 +2,32 @@
...
+This library provides instrumentation for Spring Web applications (Boot, MVC and WebFlux). It creates tracing data for 
+server requests and also client requests (`RestTemplate`, `AsyncRestTemplate` and `WebClient`).
 
 ## Use [opentracing-spring-cloud](https://github.com/opentracing-contrib/java-spring-cloud) instead
 
 As it was mentioned above this library traces only inbound/outbound HTTP requests. If you would like to 
 get automatically traced different set of technologies e.g. `spring-cloud-netflix`, JMS or even more then
 use project [opentracing-spring-cloud](https://github.com/opentracing-contrib/java-spring-cloud) instead.
 
+For reactive applications, it is especially recommended to use `reactor` tracing from
```
pavolloffay: Is reactor supported in opentracing-spring-cloud?

opentracing-spring-web-itest/webflux/src/test/java/io/opentracing/contrib/spring/webflux/webfilter/itest/webflux/WebFluxJettyMinimalIntegrationTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+        final MockSpan span = mockSpans.get(0);
+        Assert.assertEquals("GET", span.operationName());
+
+        Assert.assertEquals(8, span.tags().size());
+        Assert.assertEquals(Tags.SPAN_KIND_SERVER, span.tags().get(Tags.SPAN_KIND.getKey()));
+        Assert.assertEquals("GET", span.tags().get(Tags.HTTP_METHOD.getKey()));
+        Assert.assertEquals(TEST_REST_TEMPLATE.getRootUri() + "/", span.tags().get(Tags.HTTP_URL.getKey()));
+        Assert.assertEquals(200, span.tags().get(Tags.HTTP_STATUS.getKey()));
+        Assert.assertEquals("WebFlux/java-spring-web", span.tags().get(Tags.COMPONENT.getKey()));
```
pavolloffay: Change `java-spring-webflux` to be consistent with other component names in this repo?

opentracing-spring-web-starter/pom.xml
```diff
@@ -24,6 +24,11 @@
       <artifactId>spring-webmvc</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-webflux</artifactId>
+      <scope>provided</scope>
```
pavolloffay: I think this should be optional? We don't want users to depend on webflux

opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebClientTracingAutoConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+
+/**
+ * Instrumentation of {@link WebClient}. It also instruments {@link WebClient.Builder} and all instances created with it.
+ *
+ * @author Csaba Kos
+ */
+@Configuration
+@ConditionalOnBean(Tracer.class)
+@ConditionalOnClass(WebClient.class)
```
pavolloffay: Should we also do conditional on reactive?

pavolloffay: @ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.REACTIVE)

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/WebClientSpanDecorator.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        @Override
+        public void onError(final ClientRequest clientRequest, final Throwable throwable, final Span span) {
+            Tags.ERROR.set(span, Boolean.TRUE);
+            span.log(errorLogs(throwable));
+        }
+
+        @Override
+        public void onCancel(final ClientRequest httpRequest, final Span span) {
```
pavolloffay: When does cancel happen?

csabakos:
> @csabakos thanks for the PR! It looks good.

@pavolloffay Thank you for the review!

> I would like to see a test where a client is used to from multiple threads (assuming it supports concurrent processing) and creates requests to instrumented web endpoint (can be reactive too)
> 
> [java-spring-web/opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingAsyncRestTemplateTest.java](https://github.com/opentracing-contrib/java-spring-web/blob/5d7207a43ec4b8efd0314fb42ae8810eed43f4fa/opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingAsyncRestTemplateTest.java#L65)
> 
> Line 65 in [5d7207a](/opentracing-contrib/java-spring-web/commit/5d7207a43ec4b8efd0314fb42ae8810eed43f4fa)
> 
>  public void testMultipleRequests() throws InterruptedException, ExecutionException {

I implemented the exact test that you linked to for WebClient: [opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingWebClientTest.java](https://github.com/csabakos/java-spring-web/blob/e41a79f48c626f48e200459dc7f4c61e4cdeb051/opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingWebClientTest.java#L68). Is that sufficient or did you have something else in mind?

> I have noticed that some files were taken from spring-cloud-sleuth. I think we should add their license (We already doing it for some files in spring-cloud repository)

Conveniently their license is also Apache License 2.0, but I will gladly add it. Would you mind pointing me to an existing example of what exactly needs to be done?

README.md
csabakos: I'm planning to add that in a separate pull request. I already have the code ready for it, but `opentracing-reactor` needs to be published to Maven central first. (See https://github.com/opentracing-contrib/java-reactor/issues/1)

opentracing-spring-web-starter/pom.xml
csabakos: Thanks, `optional` makes more sense than `provided`.

opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebClientTracingAutoConfiguration.java
csabakos: `@ConditionalOnWebApplication` refers to the server technology.

One could use the reactive `WebClient` in a Servlet-based server application, so I don't think that adding that conditional would be appropriate.

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/WebClientSpanDecorator.java
csabakos: Excellent question! It looks like cancellation can happen in various circumstances:
- If the user of the WebClient "unsubscribes" from the request (for example because they have a timeout that they reached and they no longer wish to wait for the result). I think in this case I think it is correct to log an error on the span, since timeout can be considered an error.
- In various Reactor constructs, `cancel()` is called even when the `Mono` (successfully) emits an item. An example of this is [`Flux#next()`](https://github.com/reactor/reactor-core/blob/7bf86a81734bf1dc1a468a99dfdad2590c089de9/reactor-core/src/main/java/reactor/core/publisher/MonoNext.java#L75). Since this is part of the normal (successful) behavior, I don't think we should log an error.

There does not seem to be a good way to differentiate between the above two cases in the tracing code, so I changed this to log a `cancelled` event instead of an error. It will be up to the user who is looking at the trace logs to determine if the cancellation indicates an error or not. (For example the user can infer this from the HTTP status code, etc.)

pavolloffay:
> Conveniently their license is also Apache License 2.0, but I will gladly add it. Would you mind pointing me to an existing example of what exactly needs to be done?

I think just add the license to the files and then exclude them in pom https://github.com/opentracing-contrib/java-spring-cloud/blob/master/pom.xml#L380

opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebFluxTracingAutoConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+ * Instrumentation of WebFlux.
+ *
+ * @author Csaba Kos
+ */
+@Configuration
+@ConditionalOnBean(Tracer.class)
+@AutoConfigureAfter(TracerAutoConfiguration.class)
+@EnableConfigurationProperties(WebTracingProperties.class)
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.REACTIVE)
+@ConditionalOnProperty(name = "opentracing.spring.web.enabled", havingValue = "true", matchIfMissing = true)
```
pavolloffay: `CONFIGURATION_PREFIX` can be used

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/WebClientSpanDecorator.java
pavolloffay: If it is a part of normal execution flow and it happes often we can avoid adding meta-data at all.

opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingWebClientTest.java
```diff
@@ -0,0 +1,133 @@
...
+        stubFor(get(urlPathMatching(".*/foo/.*"))
+                .willReturn(aResponse().withStatus(200)));
+
+        ExecutorService executorService = Executors.newFixedThreadPool(100);
+        List<Future<?>> futures = new ArrayList<>(numberOfCalls);
+        for (int i = 0; i < numberOfCalls; i++) {
+            final String requestUrl = url + i;
+
+            final Scope parentSpan = mockTracer.buildSpan("foo").startActive(false);
+            parentSpan.span().setTag("request-url", requestUrl);
```
pavolloffay: Add tag on spanBuilder

opentracing-spring-web/src/test/java/io/opentracing/contrib/spring/web/client/TracingWebClientTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        stubFor(get(urlPathMatching(".*/foo/.*"))
+                .willReturn(aResponse().withStatus(200)));
+
+        ExecutorService executorService = Executors.newFixedThreadPool(100);
+        List<Future<?>> futures = new ArrayList<>(numberOfCalls);
+        for (int i = 0; i < numberOfCalls; i++) {
+            final String requestUrl = url + i;
+
+            final Scope parentSpan = mockTracer.buildSpan("foo").startActive(false);
```
pavolloffay: Create a span directly and remove `cont`. cont does not seems to be needed

geoand:
> @geoand could you please review too?

I will do that tomorrow :)

csabakos:
> > Conveniently their license is also Apache License 2.0, but I will gladly add it. Would you mind pointing me to an existing example of what exactly needs to be done?
> 
> I think just add the license to the files and then exclude them in pom https://github.com/opentracing-contrib/java-spring-cloud/blob/master/pom.xml#L380

Done. The pom did not have `license-maven-plugin` so I went ahead and added that and updated all of the license headers (so that the build passes).

opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/client/WebClientSpanDecorator.java
csabakos: As an application developer, it is still valuable information that the subscription was cancelled. Cancellation does not happen in simple WebClient flows, only if you're doing something more complex (like retries, or firing off multiple requests and waiting for the fastest one to return).

I think it's worth have these events in the trace logs.

If other people find them bothersome, they can file an issue and we can make it optional, but I think that won't be the case.

geoand:
@csabakos Can you please resolve the conflict? Thanks

csabakos:
> @csabakos Can you please resolve the conflict? Thanks

@geoand resolved by merging `master` into my feature branch. (Let me know if that's not the preferred way to do this.)

geoand:
That's up to @pavollofay to decide 😎

opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebClientTracingAutoConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+@ConditionalOnProperty(prefix = WebClientTracingProperties.CONFIGURATION_PREFIX, name = "enabled", matchIfMissing = true)
+@AutoConfigureAfter(TracerAutoConfiguration.class)
+@EnableConfigurationProperties(WebClientTracingProperties.class)
```
geoand: Should `WebClientTracingProperties` be renamed? I mean the name is a little misleading since it's configuration for `RestTemplate`, `AsyncRestTemplate` and `WebClient`.
Also, the comments in `WebClientTracingProperties` should be updated to reflect this fact

opentracing-spring-web-starter/pom.xml
```diff
@@ -24,6 +43,12 @@
       <artifactId>spring-webmvc</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-webflux</artifactId>
+      <scope>compile</scope>
```
geoand: This is default scope, so perhaps best to omit it?

geoand:
Looks good from a Spring perspective :+1:

opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebClientTracingAutoConfiguration.java
csabakos: I don't have an opinion about renaming - I could see arguments both pro and con. I'll leave the decision up to you and @pavolloffay. (Because the class is `public`, it is in theory part of the public API, so that'd be one reason for not renaming it.)

I will update the comments though.

geoand: It's true that it's public, public configuration classes would rarely be used directly in application code. I would assume that at least a minor version bump would be warranted even without the rename, so some minor breaking could probably be expected

opentracing-spring-web-starter/pom.xml
pavolloffay: Shouldn't this be optional?

csabakos: Nice catch! Looks like I only updated the POM in `opentracing-spring-web-starter` but not in `opentracing-spring-web`. Fixed now.

pavolloffay:
@geoand can this be merged?

geoand:
@pavolloffay yeah it LGTM. Should I merge and cut a release?

pavolloffay:
:+1:

geoand:
Thanks for the awesome work @csabakos!

csabakos:
Thanks y'all for the prompt review and good comments!

I'll work on adding the Spring Boot starter for `opentracing-reactor` to `opentracing-spring-cloud` after https://github.com/opentracing-contrib/java-reactor/issues/1 is completed.

geoand:
Thanks again @csabakos!

ghilainm:
@csabakos Seems like the filter is not instantiated if spring-mvc is not on the classpath. Is it intented? I am using spring cloud gateway and the filter is not created because of this :).

PR #96, opened 2019-03-08
WebFlux tracing should trace all URLs by default.

csabakos:
The `urlPatterns` field in `WebTracingProperties` has different semantics for servlet and reactive web server technologies. (This naturally comes from the fact that the semantics are defined in the servlet spec for the former, and by the implementation in WebFlux for the latter.)

The default value of `urlPatterns` is set to `/*`, which enables tracing of all URLs in case of servlet stack, but it means "only single path component URLs" for reactive/WebFlux (e.g. `/foo` would be traced but not `/foo/bar`).

This PR ensures that by default all URLs will be traced by both servlet and reactive stacks, and it also improves on the documentation of the `urlPatterns` field to make the difference in semantics clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ class TracingConfiguration {
                 tracer,
                 Integer.MIN_VALUE,               // Order
                 Pattern.compile(""),             // Skip pattern
-                Collections.singletonList("/*"), // URL patterns
+                Collections.emptyList(),         // URL patterns, empty list means all
                 Arrays.asList(new WebFluxSpanDecorator.StandardTags(), new WebFluxSpanDecorator.WebFluxTags())
         );
     }
--- a/opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebTracingProperties.java
+++ b/opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/WebTracingProperties.java
@@ -33,7 +33,18 @@ public class WebTracingProperties {
     private boolean enabled = true;
     private Pattern skipPattern = DEFAULT_SKIP_PATTERN;
     private int order = Integer.MIN_VALUE;
-    private List<String> urlPatterns = new ArrayList<>(Collections.singletonList("/*"));
+
+    /**
+     * List of URL patterns that should be traced.
+     *
+     * For servlet web stack, the URL pattern syntax is defined in the Servlet spec, under "Specification of Mappings".
+     * For reactive (WebFlux), see the documentation of {@link org.springframework.web.util.pattern.PathPattern} for
+     * the syntax.
+     *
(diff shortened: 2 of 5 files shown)


geoand:
@csabakos Thanks, this looks good!

Can we have a couple tests also where the `urlPatterns` are not set in properties to ensure that the new defaults work as expected?

csabakos:
> @csabakos Thanks, this looks good!
> 
> Can we have a couple tests also where the `urlPatterns` are not set in properties to ensure that the new defaults work as expected?

Hi @geoand and thanks for the review! I added `ServerTracingAutoConfigurationDefaultsTest.java` and `WebFluxTracingAutoConfigurationDefaultsTest.java` with that intention. Do you think more test cases are needed, or should I add some documentation about what those tests are trying to check?

geoand:
> > @csabakos Thanks, this looks good!
> > Can we have a couple tests also where the `urlPatterns` are not set in properties to ensure that the new defaults work as expected?
> 
> Hi @geoand and thanks for the review! I added `ServerTracingAutoConfigurationDefaultsTest.java` and `WebFluxTracingAutoConfigurationDefaultsTest.java` with that intention. Do you think more test cases are needed, or should I add some documentation about what those tests are trying to check?

I personally think that some documentation would be just fine. @pavolloffay what are your thoughts on this?

README.md
```diff
@@ -99,7 +99,7 @@ class TracingConfiguration {
                 tracer,
                 Integer.MIN_VALUE,               // Order
                 Pattern.compile(""),             // Skip pattern
-                Collections.singletonList("/*"), // URL patterns
+                Collections.emptyList(),         // URL patterns, empty list means all
```
pavolloffay: If empty means all - how one can disable webflux tracing for all URLs?

pavolloffay:
I leave it to you @geoand. I think the PR looks fine.

geoand:
I will take another look tomorrow, but my first impression is that this is good :)

README.md
csabakos: They can disable it using `opentracing.spring.web.enabled = false`.

It's worth noting that the existing behavior for the MVC `ServerTracingAutoConfiguration` is the same: it binds to all URLs if an empty `urlPattern` list is specified.

This is because that is the default behavior in [FilterRegistrationBean](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/servlet/FilterRegistrationBean.html):
> When no URL pattern or servlets are specified the filter will be associated to '/*'.

geoand:
@csabakos This LGTM! I take it that you have completed the work and we can merge?

csabakos:
@geoand Thank you! Yes, please feel free to merge any time.

geoand:
Thank you for the PR @csabakos!

Issue #103, opened 2019-05-02
Avoid circular beans creation when zuul is on the classpath


ask4gilles:
see context here https://github.com/opentracing-contrib/java-spring-cloud/issues/219

PR #104, opened 2019-05-02
avoid circular beans creations.

ask4gilles:
fix https://github.com/opentracing-contrib/java-spring-web/issues/103 and https://github.com/opentracing-contrib/java-spring-cloud/issues/219


ask4gilles:
cc @pavolloffay

PR #109, opened 2019-07-03
Ensure that unit tests can be run when WebEnvironment.NONE is used

### LINKED ISSUE #108: Unmatched dependencies between beans
ServerTracingAutoConfiguration has the following dependency:
@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
but is also responsible for loading WebTracingProperties.
SkipPatternAutoConfiguration depends on WebTracingProperties being loaded, but is not conditional on Type.SERVLET. Thus, if you run a unittest using webEnvironment = SpringBootTest.WebEnvironment.NONE, Spring will try to instantiate SkipPatternAutoConfiguration, but it will fail since WebTracingProperties is not in the ApplicationContext.

I'm sorry that I don't have a suggested solution, but I don't have the full overview of what is actually required or not in a WebEnvironment.NONE situation.

A reproducer is available at https://github.com/scoof/opentracing-contrib-java-spring-web-reproducer/

geoand:
Fixes #108

### DIFF
--- a/opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/SkipPatternAutoConfiguration.java
+++ b/opentracing-spring-web-starter/src/main/java/io/opentracing/contrib/spring/web/starter/SkipPatternAutoConfiguration.java
@@ -143,6 +143,9 @@ public class SkipPatternAutoConfiguration {
   @Configuration
   protected static class DefaultSkipPatternConfig {
 
+    @Autowired(required = false)
+    WebTracingProperties webTracingProperties;
+
     private static String combinedPatterns(String skipPattern) {
       String pattern = skipPattern;
       if (!StringUtils.hasText(skipPattern)) {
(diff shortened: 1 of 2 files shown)


geoand:
@ask4gilles care to review?

scoof:
I'm not sure this is a proper fix. Wouldn't this ignore WebTracingProperties even if the properties were present, when running under WebEnvironment.NONE?

geoand:
Yes, indeed it would. 

Honest question: why would need to set these properties and have all machinery populated when using `WebEnvironment.NONE`?

scoof:
Fair question. However, I don't have enough of an overview to answer that question. I can't say if this will be usable in that context.

geoand:
Me neither :)

Lets see if @ask4gilles has info

ask4gilles:
@geoand I’m on vacation until next week ;)

geoand:
Have fun @ask4gilles !

I'll merge in the meantime and we can reexamine in the future if needed

PR #110, opened 2019-07-05
add uber-trace-id for request header

aslongyouloveme:
so microservice can get this for inherit span

### DIFF
--- a/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingOperator.java
+++ b/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingOperator.java
@@ -20,6 +20,7 @@ import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
 import io.opentracing.propagation.Format;
+import io.opentracing.propagation.TextMapAdapter;
 import io.opentracing.tag.Tags;
 import org.springframework.http.server.reactive.ServerHttpRequest;
 import org.springframework.web.server.ServerWebExchange;
@@ -28,7 +29,10 @@ import reactor.core.publisher.Mono;
 import reactor.core.publisher.MonoOperator;
 import reactor.util.context.Context;
 
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Similar to {@code MonoWebFilterTrace} from spring-cloud-sleuth-core.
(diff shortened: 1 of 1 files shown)


opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingOperator.java
```diff
@@ -72,7 +76,21 @@ public void subscribe(final CoreSubscriber<? super Void> subscriber) {
...
         }
     }
+
+    private ServerWebExchange wrappleExchange(Span span) {
+        Map<String, String> map = new HashMap();
+        this.tracer.inject(span.context(), Format.Builtin.HTTP_HEADERS, new TextMapAdapter(map));
+        ServerHttpRequest req = this.exchange.getRequest();
+        Iterator iterator = map.keySet().iterator();
+
+        while (iterator.hasNext()) {
```
williamokano: Since this is an iterator, would not be better using a "simple" `for (String key : map.keySet())`? This way you don't have to manually call `iterator.next()`.

williamokano: Since this is an iterator, would not be better using a "simple" `for (String key : map.keySet())`? This way you don't have to manually call `iterator.next()`.

aslongyouloveme: yes, thank you

pavolloffay:
@aslongyouloveme could you please help me understand this change and describe why it is needed? 

The request already contains span context. This span context was used as a parent to span created in the subscribe method.

aslongyouloveme:
I used springcloud-gateway as the gateway, and downstream order-to-service is a micro service, gateway is a webflux application, and order-to-service is a servlet application. When I integrated jaeger, order-to-service could not get spancontext of gateway, because header (ude-trace-id) was not added in the forwarded request header. @

aslongyouloveme:
@pavolloffay

pavolloffay:
It sounds fine, thanks for the explanation. Could you write a test for the functionality you are adding?

aslongyouloveme:
I would like to write a test case for it, I did not find any specimen test case for ‘TracingOperator.java’. 
It seams difficult to mock ServerWebExchange that has so much method, 
e.g this.exchange.mutate().request(req).build(), req.mutate().header(key, map.get(key))
Could you please give me some help?

pavolloffay:
I am more interested in integration test. There are some for webflux which should use this class.

Issue #113, opened 2019-08-12
SkipPattern for Management Endpoints doesn't work with context-path


ancore:
The skip pattern for actuator endpoints contains the context-path, but the TracingFilter tries to match the skip pattern without context-path.

The following is added to the configured default skip pattern:
`/user/api/management/(health|health/.*|info|info/.*)`

The method `TracingFilter#isTraced` removes the context-path from the request URI:
`httpServletRequest.getRequestURI().substring(httpServletRequest.getContextPath().length())` 

Maybe I get something wrong, but I suspect that this filter never works when Spring has a context-path configured.

pavolloffay:
cc)  @larchanjo @ask4gilles  could you please have a look. You worked on the actuator part.


Does the context path affect the actuator endpoints? 

I think we should remove the context path from the actuator skip pattern and add e2e test with configured context-path.

Would you be interested in submitting a PR?

ancore:
I'm trying to understand the unit tests and found this while debugging SkipEndPointsWithContextPathAndBasePathTest#testTraceHealthCareEndpoint:

![istTraced](https://user-images.githubusercontent.com/1880335/62860697-53cbe880-bd01-11e9-8656-af6a83774a1d.png)

ancore:
The TestRestTemplate expands the URL automatically by context-path (RestTemplate#execute).
When you prefix your URL to call for tests with the context-path
`private static final String HEALTHCARE = CONTEXT_PATH + "/healthcare";`
it will be there twice and therefore can be removed once.

I think, the unit test is wrong and hides the actual error I mentioned in my initial post.

I would not like to submit a PR because it affects a lot of unit test methods and I can't estimate the impact.

ask4gilles:
@pavolloffay @ancore I'll have a look at this tomorrow.

ask4gilles:
@ancore In order to gain time, can you please provide a minimal project sample to reproduce your issue? Thanks!

ancore:
You will see the effect when you remove `CONTEXT_PATH +` from
`private static final String HEALTHCARE = CONTEXT_PATH + "/healthcare"`

ask4gilles:
@ancore What's "the effect"? 
If I removed it, I still see it traced, which is normal.
What's your issue exactly? An actuator endpoint traced which should not or the other way around?
When you say:

> The following is added to the configured default skip pattern:
/user/api/management/(health|health/.*|info|info/.*)

[...]

PR #114, opened 2019-08-15
remove context-path from skip pattern configuration.

ask4gilles:
cc @pavolloffay 
Fix https://github.com/opentracing-contrib/java-spring-web/issues/113


opentracing-spring-web-starter/src/test/java/io/opentracing/contrib/spring/web/starter/SkipEndPointsWithContextPathWithoutBasePathTest.java
```diff
@@ -51,16 +51,15 @@
 @RunWith(SpringJUnit4ClassRunner.class)
 public class SkipEndPointsWithContextPathWithoutBasePathTest {
 
-  private static final String CONTEXT_PATH = "/myapp";
```
pavolloffay: Is there a test which sets a context path?

ask4gilles: @pavolloffay As pointed in the issue, since the TestRestTemplate always expands the context-path since `server.servlet.context-path:/myapp` all tests endpoints contain the context-path
snap-4

PR #126, opened 2020-01-21
increase the visibility of shouldFilter.

sijuv:
There could be need to further customize the default shouldFilter implementation, ex you want to exclude traffic based on user agent as in case of bots.

### DIFF
--- a/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingWebFilter.java
+++ b/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingWebFilter.java
@@ -98,7 +98,7 @@ public class TracingWebFilter implements WebFilter, Ordered {
      *
      * @return whether request should be traced or not
      */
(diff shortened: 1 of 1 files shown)


geoand:
Thanks for the contribution.

Can you show a concrete example where you would need this?

Thanks

sijuv:
we have a use case where certain bots hit our production services. These are identified by their user-agent header. We dont want such calls to be traced. 
A sample code would look like this:

    if (Option.ofNullable(request.getHeaders().getFirst(HttpHeaders.USER_AGENT))
        .map(s -> userAgentSkipPattern.matcher(s.toLowerCase()).matches()).orElse(false)) {
      if (LOG.isTraceEnabled()) {
        LOG.trace(
            "Not tracing request " + request + " because it matches user agen skip pattern: " + userAgentSkipPattern);
      }
      return false;
    }

To implement this I would extend the open tracing provided TracingWebFIlter and override the shouldBeTracedMethod.

geoand:
That makes sense, thanks

geoand:
I merged and launched a release of 3.0.1 which should contain the fix if the release goes well

PR #137, opened 2020-10-14
Include general enable property for RestTemplateTracingAutoConfiguration

huehnerlady:
I would expect, if you set the property `opentracing.spring.web.enabled` to false, that everything belonging to it will be disabled as well.
The RestTemplateTracingAutoConfiguration currently just listens to the property `opentracing.spring.web.client.enabled`, which is a sub property of the other one.
Therefore I suggest to include the basic property in the Conditional and just enable this configuration if both are enabled or missing (opt-out)


geoand:
Thanks for this!

I am on the fence about it to be honest. Let's see what others think

huehnerlady:
@geoand any update on what people think? Where can I find a discussion about it (if there is any?)

geoand:
No comments so far unfortunately

huehnerlady:
If I do not here anything by the end of the year, I would close this pull request, as I am not a fan of loose ends :)

huehnerlady:
Ok as it seems not to get anywhere I will close this PR now

PR #115, opened 2019-08-16
Add Spring web operation name handler

zouyx:
1.Support Spring RequestMapping
2.HandlerUtils.requestMapping，if use ```.path()```can not get anything when use ```@Requestmapping```
so change to value()

PR #139, opened 2020-12-14
Add Spring web operation name handler

zouyx:
1.Support Spring RequestMapping
2.HandlerUtils.requestMapping，if use ```.path()```can not get anything when use ```@Requestmapping```
so change to value()

### DIFF
--- a/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/HandlerInterceptorSpanDecorator.java
+++ b/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/interceptor/HandlerInterceptorSpanDecorator.java
@@ -133,6 +133,34 @@ public interface HandlerInterceptorSpanDecorator {
     };
 
     /**
+     * Decorator to record details about the handler as log events recorded on the span.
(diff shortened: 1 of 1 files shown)


geoand:
Hi,

Thanks for the contribution.

Can you please explain the rationale behind these changes?

zouyx:
> Hi,
> 
> Thanks for the contribution.
> 
> Can you please explain the rationale behind these changes?

Relate to #115 which can not run travis CI.

Just recreate it.

geoand:
I see thanks.

Could you please add a test as was requested in the original PR?

zouyx:
> I see thanks.
> 
> Could you please add a test as was requested in the original PR?

I think will add it in this few day

zeitlinger:
@zouyx not sure I understand this PR correctly. Is this about changing the operation name to be something more meaningful than `GET`?

zouyx:
> @zouyx not sure I understand this PR correctly. Is this about changing the operation name to be something more meaningful than `GET`?

yes, you are right. because it can not get anything when use @Requestmapping

zeitlinger:
Fixes #143

Issue #140, opened 2020-12-14
Span customization - Extend WebFluxSpanDecorator


NHebrard:
Hello!

I need to customize a span right after its creation and before its activation (injecting some baggage items).

I am encountering this need for both regular Servlet based applications and WebFlux applications.

In the Servlet world, it’s totally feasible thanks to a ` ServletFilterSpanDecorator.onRequest()` which is invoked right after the span creation and before its activation.
See [TracingFilter](https://github.com/opentracing-contrib/java-web-servlet-filter/blob/master/opentracing-web-servlet-filter/src/main/java/io/opentracing/contrib/web/servlet/filter/TracingFilter.java#L184).

However, it’s not achievable in the reactor world as the `WebFluxSpanDecorator.onRequest()` is actually called at the subscription time occurring after the span activation.
See [TracingOperator](https://github.com/opentracing-contrib/java-spring-web/blob/master/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingOperator.java#L73) and [TracingSubscriber](https://github.com/opentracing-contrib/java-spring-web/blob/e394c57455bd60ea05a325a1417152c0b8386f56/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/TracingSubscriber.java#L61).

Could we extend the `WebFluxSpanDecorator` by adding a new callback `onCreate()` that would be called right after the span creation?
I would also remove `This is called right after span in created` from the `onRequest()` [documentation](https://github.com/opentracing-contrib/java-spring-web/blob/e394c57455bd60ea05a325a1417152c0b8386f56/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/WebFluxSpanDecorator.java#L43) as it's not really the case.

Let me know if you agree with the proposal, I'll be happy to contribute! :)

geoand:
Hi,

That sounds good to me! Feel free to contribute :)

PR #141, opened 2021-01-04
Add onCreate callback to WebFluxSpanDecorator

NHebrard:
Hello!

With this PR, `WebFluxSpanDecorator` will be extended with an `onCreate()` callback called right after the span creation.
Cf this issue: #140

Thanks, @geoand.

### DIFF
--- a/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/WebFluxSpanDecorator.java
+++ b/opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/WebFluxSpanDecorator.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016-2019 The OpenTracing Authors
+ * Copyright 2016-2021 The OpenTracing Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


opentracing-spring-web/src/main/java/io/opentracing/contrib/spring/web/webfilter/WebFluxSpanDecorator.java
```diff
@@ -34,8 +34,17 @@
...
     /**
-     * Decorate span before {@code .onSubscribe()} is called. This is called right after span in created. Span
-     * context is already present in request attributes with name {@link TracingWebFilter#SERVER_SPAN_CONTEXT}.
+     * Decorate span right after its creation.
+     * Span context is already present in request attributes with name {@link TracingWebFilter#SERVER_SPAN_CONTEXT}.
+     *
+     * @param exchange web exchange
+     * @param span span to decorate
+     */
+    void onCreate(ServerWebExchange exchange, Span span);
```
geoand: Best make this a `default` method that is a no-op

geoand:
@csabakos wanna take a look as well?

csabakos:
> @csabakos wanna take a look as well?

The change LGTM but it would be good to understand a bit more about the use case to be able to say whether the change is really necessary or not. For example, why isn't the baggage getting extracted from the incoming HTTP headers or the parent span?

kkrawczy:
@csabakos 
The use case would be to add to the baggage custom items that are not present in the incoming HTTP headers.
WebFluxSpanDecorator sounds like a good component that could do it, but there is one problem.
During span activation, we put baggage values into MDC, so if WebFluxSpanDecorator operates after activation we miss the values that we want to inject.

